PR #4, opened 2024-08-09
CVSL-2068: Add jpa config, run script and Intellij run configuration.…

justin-gardiner:
… Add db to docker-compose. Add initial flyway script for assessment table.


gurps-singh (dismissed):

src/main/resources/migration/common/V10__create_assessment_table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE assessment
+(
+    id SERIAL NOT NULL constraint assessment_pk PRIMARY KEY,
+    booking_id integer NOT NULL,
+    prisoner_number varchar(7) NOT NULL,
+    created_timestamp TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    last_updated_timestamp TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP
```
andrewrlee, line 7: 🥇

.run/HmppsAssessForEarlyReleaseApi.run.xml
```diff
@@ -0,0 +1,29 @@
...
+      <option name="IS_ENABLE_EXPERIMENTAL_INTEGRATIONS" value="false" />
+      <ENTRIES>
+        <ENTRY IS_ENABLED="true" PARSER="runconfig" IS_EXECUTABLE="false" />
+      </ENTRIES>
+    </extension>
+    <method v="2">
+      <option name="Make" enabled="true" />
+    </method>
+  </configuration>
+</component>
```
andrewrlee, line 29: we might not be able to keep this after we hook in system creds.. unless there's a way to reference secrets

src/test/resources/application-test.yml
```diff
@@ -1,3 +1,35 @@
+spring:
+  security:
+    oauth2:
+      resourceserver:
+        jwt:
+          public-key-location: classpath:local-public-key.pub
+
+  datasource:
+    url: 'jdbc:h2:mem:afer-db;MODE=PostgreSQL'
```
andrewrlee, line 9: we might want to run tests against a real postgres server eventually but this makes a sensible default to start with 👍

gurps-singh (approved):

.run/HmppsAssessForEarlyReleaseApi.run.xml
justin-gardiner: ok no problem, just thought it would make it easier for other devs to get it running in Intellij

src/main/resources/application.yml
```diff
@@ -23,6 +23,34 @@ spring:
         provider:
           hmpps-auth:
             token-uri: ${hmpps-auth.url}/oauth/token
+  jpa:
+    open-in-view: false
+    properties:
+      hibernate:
+        dialect: org.hibernate.dialect.PostgreSQLDialect
```
andrewrlee, line 30: this doesn't produce an error on start up does it? 

We get a warning in the logs for CVL around not having to specify the dialect

andrewrlee (approved):
SWEET

LGTM

src/main/resources/application.yml
justin-gardiner: yes, I can see it,
HHH90000025: PostgreSQLDialect does not need to be specified explicitly using 'hibernate.dialect' (remove the property setting and it will be selected by default) 

I'll remove in my next pr

PR #9, opened 2024-08-22
 CVSL-2071-add-prison-id-column-to-assessment

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Assessment.kt
```diff
@@ -20,6 +20,8 @@ data class Assessment(
 
   val prisonerNumber: String,
 
+  val prisonId: String? = null,
```
justin-gardiner, line 23: Can this be non-nullable?

nishanthmahasamudram: updated to non-nullable.

src/main/resources/migration/common/V11__add_prison_id_column_to_assessment_table.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE assessment ADD COLUMN prison_id VARCHAR(3) NOT NULL;
```
justin-gardiner, line 1: I didn't think this would work because you are adding a new not null column without a default value but it looks it does. I guess it's because we don't have any entries.

justin-gardiner (approved):
LGTM

PR #10, opened 2024-08-23
CVSL-2090: Rename Assessment table to Offender and add some new columns.

justin-gardiner:


src/main/resources/migration/common/V12__rename_assessment_to_offender.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE assessment RENAME TO offender;
+
+ALTER TABLE offender ADD COLUMN first_name VARCHAR(35);
+ALTER TABLE offender ADD COLUMN last_name VARCHAR(35);
+ALTER TABLE offender ADD COLUMN hdced DATE;
```
nishanthmahasamudram, line 5: Can we add not null?

justin-gardiner: Yes we can because we only create records for offenders with an HDCED, updated

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/repository/OffenderRepository.kt
```diff
@@ -2,7 +2,9 @@ package uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.repository
 
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.stereotype.Repository
-import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.Assessment
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.Offender
 
 @Repository
-interface AssessmentRepository : JpaRepository<Assessment, Long>
+interface OffenderRepository : JpaRepository<Offender, Long> {
+  fun findByPrisonerNumber(prisonerNumber: String): Offender?
```
nishanthmahasamudram, line 9: Think it should return list of Offenders?

justin-gardiner: At the moment we are only going to have one entry for each prisonerNumber so it will only return one at the moment.

nishanthmahasamudram (approved):
LGTM

PR #11, opened 2024-08-27
Cvsl 2071 afer add consumption of domain events to capture when a prisoner is transferred

andrewrlee:


nishanthmahasamudram:
s

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Offender.kt
```diff
@@ -25,7 +25,7 @@ data class Offender(
   @NotNull
   val prisonerNumber: String,
 
-  val prisonId: String,
+  var prisonId: String,
```
justin-gardiner, line 28: Does this need to be var? If you need to update it for an existing offender, could you instead do

val updatedOffender = offender.copy(prisonId = newPrisonId)

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/SampleTest.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi
```
justin-gardiner, line 1: Is this still needed?

nishanthmahasamudram: Yes, it's just to fix gradlew test failure.

justin-gardiner (approved):
LGTM

PR #24, opened 2024-09-19
CVSL-2155 adding some logic to determine overall state

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/PolicyVersion.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.model
+
+enum class PolicyVersion(val version: String) {
```
andrewrlee, line 3: This isn't used yet, I'll do that in a follow up

justin-gardiner (approved):
LGTM

PR #31, opened 2024-09-30
CVSL-2160: Add new tables and entity classes for address check requests.

justin-gardiner:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/AddressServiceIntTest.kt
```diff
@@ -83,6 +95,54 @@ class AddressServiceTest : SqsIntegrationTestBase() {
     assertThat(savedAddress?.country).isEqualTo("England")
   }
 
+  @Sql(
```
justin-gardiner, line 98: Will update these with tests that test the service layer when it is done. They were useful for now as they found a few bugs in the db script and mapping.

detekt-baseline.xml
```diff
@@ -6,6 +6,8 @@
     <ID>ConstructorParameterNaming:PrisonOffenderEventListener.kt$HMPPSEventType$val Value: String</ID>
     <ID>ConstructorParameterNaming:PrisonOffenderEventListener.kt$HMPPSMessage$val Message: String</ID>
     <ID>ConstructorParameterNaming:PrisonOffenderEventListener.kt$HMPPSMessage$val MessageAttributes: HMPPSMessageAttributes</ID>
+    <ID>LongParameterList:CasCheckRequest.kt$CasCheckRequest$( id: Long = -1, caAdditionalInfo: String?, ppAdditionalInfo: String?, dateRequested: LocalDate = LocalDate.now(), preferencePriority: AddressPreferencePriority, status: AddressCheckRequestStatus = AddressCheckRequestStatus.IN_PROGRESS, @ManyToOne @JoinColumn(name = "allocated_address_id", referencedColumnName = "id") val allocatedAddress: Address? = null, )</ID>
```
justin-gardiner, line 9: These were actually ok, it seems to be a problem with detekt, it seems to treat annotations like @ManyToOne and @JoinColumn in the constructor as parameters.

detekt-baseline.xml
```diff
@@ -6,6 +6,9 @@
...
+    <ID>LongParameterList:CurfewAddressCheckRequest.kt$CurfewAddressCheckRequest$( @Id @GeneratedValue(strategy = GenerationType.IDENTITY) @NotNull val id: Long = -1, val caAdditionalInfo: String? = null, val ppAdditionalInfo: String? = null, @NotNull val dateRequested: LocalDate, @NotNull @Enumerated(EnumType.STRING) val preferencePriority: AddressPreferencePriority, @NotNull @Enumerated(EnumType.STRING) val status: AddressCheckRequestStatus, @NotNull val createdTimestamp: LocalDateTime = LocalDateTime.now(), @NotNull val lastUpdatedTimestamp: LocalDateTime = LocalDateTime.now(), )</ID>
```
justin-gardiner, line 10: These two were actually ok, it seems to be a problem with detekt, it seems to treat annotations like https://github.com/manytoone and https://github.com/joincolumn in the constructor as parameters, but only in the subclasses.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/repository/StandardAddressCheckRequestRepository.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.StandardAddressCheckRequest
+
+interface StandardAddressCheckRequestRepository : JpaRepository<StandardAddressCheckRequest, Long>
```
andrewrlee, line 6: you can share the same repository for both and return different sub classes.. but maybe it makes sense to keep them separate?

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/repository/StandardAddressCheckRequestRepository.kt
justin-gardiner: Cool didn't think about that, I look at adding that in my next PR. It may make sense as I can't see them having too many methods but some might get duplicated

PR #32, opened 2024-10-01
CVSL-2160: Add endpoints to add new standard address and CAS check re…

justin-gardiner:
…quests.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/resource/OffenderResource.kt
```diff
@@ -171,4 +179,98 @@ class OffenderResource(private val offenderService: OffenderService) {
...
+  )
+  @ApiResponses(
+    value = [
+      ApiResponse(
+        responseCode = "201",
+        description = "The CAS check request has been added to the current assessment for the offender.",
+        content = [
+          Content(
+            mediaType = "application/json",
+            schema = Schema(implementation = StandardAddressCheckRequestSummary::class),
```
andrewrlee, line 246: should this be CAS?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/resource/OffenderResource.kt
```diff
@@ -171,4 +179,98 @@ class OffenderResource(private val offenderService: OffenderService) {
 
     offenderService.optOut(prisonNumber)
   }
+
+  @PostMapping("/offender/{prisonNumber}/current-assessment/standard-address-check-request")
```
andrewrlee, line 183: Can we put these on an address controller? 

This this controller will get massive otherwise

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/resource/OffenderResource.kt
justin-gardiner: yes :) will fix in a follow up

PR #40, opened 2024-10-07
CVSL-2155 Fixing criterion met logic and renaming table

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/EligibilityAndSuitabilityService.kt
```diff
@@ -99,32 +99,32 @@ class EligibilityAndSuitabilityService(
...
+    val currentResults = assessment.eligibilityCheckResults
+    val criterion = policyService.getCriterion(assessment.policyVersion, criterionType, answer.code)
+    val existingCriterionResult = currentResults.find(criterionType, answer)
 
     val criteria = when {
-      existingCriteria != null -> {
-        criteriaChecks.remove(existingCriteria)
-        existingCriteria.copy(
-          criteriaMet = answer.answers.all { it.value },
-          questionAnswers = answer.answers.toMap(HashMap()),
```
andrewrlee, line 113: this was odd and unnecessary

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/policy/model/Criterion.kt
```diff
@@ -1,15 +1,26 @@
 package uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.service.policy.model
 
+fun interface EvaluationStrategy {
+  fun isMet(answers: Map<String, Boolean>): Boolean
+
+  companion object {
+    val MET_IF_ALL_ARE_FALSE = EvaluationStrategy { answers -> answers.values.all { !it } }
```
andrewrlee, line 7: this is the default..
swapped the logic round too - it used to be eligible if all were answered true

src/main/resources/migration/common/V23__drop_criteria_check_add_criterion_result.sql
```diff
@@ -0,0 +1,14 @@
+DROP TABLE criteria_check;
```
andrewrlee, line 1: Just dropped and recreated as we don't have any data we need to keep in any envs

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/policy/model/Criterion.kt
```diff
@@ -1,15 +1,26 @@
...
+  }
+}
+
 data class Criterion(
   val code: String,
   val name: String,
   val questions: List<Question> = emptyList(),
-) {
+  val evaluationStrategy: EvaluationStrategy = EvaluationStrategy.MET_IF_ALL_ARE_FALSE,
+) : EvaluationStrategy by evaluationStrategy {
```
andrewrlee, line 17: think this is the first time I've used kotlin delegation

andrewrlee: rather than having to say that Criterion implements evaluation strategy and then 

`fun isMet(questions: Map<string, boolean>) = evaluationStrategy.isMet(questions)`

louisenorris (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/policy/model/Criterion.kt
justin-gardiner: yeah I don't think I've ever seen it used outside of a book. Sure it's useful once you get used to it

justin-gardiner (commented):
LGTM

PR #41, opened 2024-10-08
CVSL-2175: Update add standard address check request to add the addre…

justin-gardiner:
…ss if it doesn't exist. Update getAddressesForPostcode to return 401 instead of 500 for an invalid postcode. Fix address first line to include organisation name if one is populated.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/os/OsPlacesApiClient.kt
```diff
@@ -18,6 +21,7 @@ class OsPlacesApiClient(
       .accept(MediaType.APPLICATION_JSON)
       .retrieve()
       .bodyToMono(OsPlacesApiResponse::class.java)
+      .onErrorResume { e -> if (e is WebClientResponseException && e.statusCode == HttpStatus.BAD_REQUEST) Mono.empty() else Mono.error(e) }
```
justin-gardiner, line 24: Handle BAD_REQUEST by returning empty list if an invalid post code format is given

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/os/OsPlacesApiResponse.kt
```diff
@@ -1,5 +1,5 @@
 package uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.service.os
 
 data class OsPlacesApiResponse(
-  val results: List<OsPlacesApiAddress>,
+  val results: List<OsPlacesApiAddress>?,
```
justin-gardiner, line 4: Normally it returns an empty array for no results found but I found one case where it returned nothing strangely, this allows us to handle that case

andrewrlee (approved):
LGTM

PR #54, opened 2024-11-06
CVSL-2200 Correcting overall eligibility status calculation

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/EligibilityAndSuitabilityService.kt
```diff
@@ -33,13 +38,24 @@ class EligibilityAndSuitabilityService(
...
       overallStatus = assessment.calculateAggregateEligibilityStatus(),
-      eligibility = assessment.eligibilityProgress(),
-      eligibilityStatus = assessment.eligibilityProgress().toStatus(),
-      suitability = assessment.suitabilityProgress(),
-      suitabilityStatus = assessment.suitabilityProgress().toStatus(),
+      eligibility = eligibility,
+      eligibilityStatus = eligibilityStatus,
+      suitability = suitability,
+      suitabilityStatus = suitabilityStatus,
+      failureType = when {
```
justin-gardiner, line 53: Will this logic have to be done everywhere we construct one of these or can it be moved into the constructor?

justin-gardiner (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/EligibilityAndSuitabilityService.kt
andrewrlee: That's a really good question - The use of extension functions make this a bit misleading.

I tried to refactor this previously but the statuses/progress are currently dynamically generated based on having access to the policy info. 

I think maybe eligibility status checks should be persisted rather than dynamically generated

PR #58, opened 2024-11-08
Cvsl 2252 afer com residential checks get com user details

nishanthmahasamudram:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/StaffResourceIntTest.kt
```diff
@@ -0,0 +1,93 @@
...
+        .exchange()
+        .expectStatus()
+        .isOk
+        .expectBody(object : ParameterizedTypeReference<User>() {})
+        .returnResult().responseBody!!
+
+      Assertions.assertThat(comUserDetails).isEqualTo(comUser)
+    }
+  }
+
```
SebNegahbanMoJ, line 66: **suggestion:** I think it may be worth adding a test to ensure it handles returning an empty array

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/probation/DeliusApiClient.kt
```diff
@@ -30,4 +30,17 @@ class DeliusApiClient(@Qualifier("oauthDeliusApiClient") val communityApiClient:
...
+    val communityApiResponse = communityApiClient
+      .post()
+      .uri("/staff")
+      .accept(MediaType.APPLICATION_JSON)
+      .bodyValue(username)
+      .retrieve()
+      .bodyToMono(typeReference<User>())
+      .block()
+    return communityApiResponse
+      ?: error("Unexpected null response from API")
```
SebNegahbanMoJ, line 44: **question:** I assume this endpoint must return an empty array if there are no staff members found?

andrewrlee: it returns a single user or will 404

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/StaffResourceIntTest.kt
andrewrlee: worth static importing Assertions, but a test to cover the 404 would be good.

nishanthmahasamudram: done.

andrewrlee (approved):
LGTM

PR #59, opened 2024-11-12
CVSL-2255: Add endpoint to retrieve COM caseload.

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/Functions.kt
```diff
@@ -3,3 +3,16 @@ package uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.service
 import org.springframework.core.ParameterizedTypeReference
 
 inline fun <reified T> typeReference() = object : ParameterizedTypeReference<T>() {}
+
```
justin-gardiner, line 6: All these were just copied from CVL. I can't see anyway of sharing stuff like this between projects at the moment. It looks like kotlin-lib is just for spring stuff etc.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/OffenderServiceTest.kt
```diff
@@ -35,13 +41,51 @@ class OffenderServiceTest {
...
     OffenderService(
+      assessmentRepository,
       offenderRepository,
       prisonerSearchService,
       probationService,
       staffRepository,
       telemetryClient,
     )
 
+  @Test
```
justin-gardiner, line 52: I noticed this was missing a unit test

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/Functions.kt
andrewrlee: not super worried about duplication for things like this..

andrewrlee (approved):
LGTM

PR #63, opened 2024-11-21
Moving over to use a state machine

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Assessment.kt
```diff
@@ -29,8 +33,12 @@ data class Assessment(
...
 
   @NotNull
-  @Enumerated(EnumType.STRING)
-  var status: AssessmentStatus = AssessmentStatus.NOT_STARTED,
+  @Embedded
+  @AttributeOverrides(
+    AttributeOverride(name = "status", column = Column(name = "status")),
+    AttributeOverride(name = "previousStatus", column = Column(name = "previous_status")),
+  )
+  var status: AssessmentState = AssessmentState.NotStarted,
```
andrewrlee, line 41: the idea would be rather than storing a enum against the record we'd stored an embeddable State object - which consist of an enum and possibly extra fields that are required to reconstitute the state.

So like if someone opts out, the opt out state also includes info about where it was opted out from (previous stage) - likely to need other things for time out

It'd be difficult to support when you have multiple levels of states you need to return from:
 ELIGIBLE -> OPT OUT -> TIMED OUT -> OPT OUT -> ELIGIBLE 

We have a history of assessment events we could possibly look at using.. 


These would be columns on the Assessment table

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Assessment.kt
```diff
@@ -89,16 +97,32 @@ data class Assessment(
...
+        log.info("Transitioned Assessment: '${this.id}', triggered by event: '${transition.event.label()}' from '${transition.fromState.label()}' to '${transition.toState.label()}'")
+        if (currentStatus != transition.fromState) {
+          assessmentEvents.add(
+            StatusChangedEvent(
+              assessment = this,
+              changes = StatusChange(before = transition.fromState.label, after = transition.toState.label),
+            ),
+          )
+          status = transition.toState
+          lastUpdatedTimestamp = LocalDateTime.now()
```
andrewrlee, line 122: transitioning adds to history and marks last updated.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Assessment.kt
```diff
@@ -89,16 +97,32 @@ data class Assessment(
...
+  private fun Any.label() = this::class.simpleName
+
+  fun performTransition(
+    event: AssessmentLifecycleEvent,
+  ): AssessmentState {
+    val currentStatus = this.status
+    val transition = assessmentStateMachine.with { initialState(currentStatus) }.transition(event)
+    return when (transition) {
+      is Invalid -> {
+        error("Fail to transition Assessment: '${this.id}', triggered by '${transition.event.label()}' from '${transition.fromState.label()}'")
```
andrewrlee, line 109: maybe this should return something rather than erroring? use cases may become apparent

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AssessmentStatus.kt
```diff
@@ -18,11 +20,6 @@ import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.UserRol
 
 enum class AssessmentStatus {
   NOT_STARTED {
-    override fun transitions() = transitions(
```
andrewrlee, line 21: transitions are part of the statemachine now

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AssessmentStatus.kt
```diff
@@ -285,16 +138,436 @@ enum class AssessmentStatus {
...
+  object TimedOut : AssessmentState {
+    override val label = AssessmentStatus.TIMED_OUT
+  }
+
+  object Postponed : AssessmentState {
+    override val label = AssessmentStatus.POSTPONED
+  }
+
+  data class OptedOut(val previousState: AssessmentState) : AssessmentState {
+    override val label = AssessmentStatus.OPTED_OUT
```
andrewrlee, line 213: Stages can have all sorts of state, but it needs to be persisted so the stage can be reconstituted

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Assessment.kt
andrewrlee: this isn't currently working - relies on persisting instances of a sealed class as an embedded entity.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AssessmentStatus.kt
```diff
@@ -285,16 +138,436 @@ enum class AssessmentStatus {
...
+      transitionTo(AssessmentState.AddressUnsuitable)
+    }
+    on<AssessmentLifecycleEvent.Refuse> {
+      transitionTo(AssessmentState.Refused)
+    }
+    on<AssessmentLifecycleEvent.Timeout> {
+      transitionTo(AssessmentState.TimedOut)
+    }
+    on<AssessmentLifecycleEvent.OptBackIn> {
+      transitionTo(this.previousState)
```
andrewrlee, line 565: so when opting out, return to the previous state that's associated with the `OptOutState`

at this level - `this` refers to the surrounding state, `it` refers to attributes of the event

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/EligibilityAndSuitabilityService.kt
```diff
@@ -102,7 +105,13 @@ class EligibilityAndSuitabilityService(
         criterionMet,
         answer.answers,
       )
-      assessmentEntity.changeStatus(assessmentLifecycleService.eligibilityAnswerSubmitted(currentAssessment))
+      val event = when (currentAssessment.calculateAggregateEligibilityStatus()) {
+        ELIGIBLE -> AssessmentLifecycleEvent.CompleteEligibilityAndSuitability
+        INELIGIBLE -> AssessmentLifecycleEvent.FailEligibilityAndSuitability
+        IN_PROGRESS -> AssessmentLifecycleEvent.StartEligibilityAndSuitability
+        NOT_STARTED -> error("Assessment: ${assessmentEntity.id} is in an unexpected state: ${assessmentEntity.status}")
```
andrewrlee, line 112: It's possible that `eligibilitystatus` should be an attribute of a new `ElgibilityAnswerSupplied` event and then the state transition could determine where to go in the statemachine:
```
  state<AssessmentState.ElgibilityAnswerSupplied> {
    on<AssessmentLifecycleEvent.CompleteEligibilityAndSuitability> {
      val newStage = when(it.eligiblityStatus) {
        ELIGIBLE -> AssessmentLifecycleEvent.CompleteEligibilityAndSuitability
        INELIGIBLE -> AssessmentLifecycleEvent.FailEligibilityAndSuitability
        IN_PROGRESS -> AssessmentLifecycleEvent.StartEligibilityAndSuitability
        NOT_STARTED -> error("Assessment: ${assessmentEntity.id} is in an 
      }
      transitionTo(newStage)
    }
```

There's lots of different states that this event could be applied to so we'd need to look at reusing this kind of logic and applying it to multiple stages

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentService.kt
```diff
@@ -67,41 +68,32 @@ class AssessmentService(
...
 
   @Transactional
   fun optOut(prisonNumber: String) {
-    val assessmentWithEligibilityProgress = getCurrentAssessment(prisonNumber)
-    val newStatus = assessmentLifecycleService.optOut(assessmentWithEligibilityProgress)
-    val assessmentEntity = assessmentWithEligibilityProgress.assessmentEntity
-
-    assessmentEntity.changeStatus(newStatus)
+    val assessmentEntity = getCurrentAssessment(prisonNumber).assessmentEntity
+    assessmentEntity.performTransition(OptOut)
```
andrewrlee, line 82: this is loads simpler

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AssessmentStatus.kt
```diff
@@ -285,16 +138,436 @@ enum class AssessmentStatus {
...
+      transitionTo(AssessmentState.PassedPreReleaseChecks)
+    }
+    on<AssessmentLifecycleEvent.FailAddressChecks> {
+      transitionTo(AssessmentState.AddressUnsuitable)
+    }
+    on<AssessmentLifecycleEvent.Refuse> {
+      transitionTo(AssessmentState.Refused)
+    }
+    on<AssessmentLifecycleEvent.OptOut> {
+      transitionTo(AssessmentState.OptedOut(this))
```
andrewrlee, line 466: so if `Ineligible/Unsuitable`, if the user selects to `optOut` then move to `OptedOut`and store the previous stage as `Ineligible/Unsuitable`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentLifecycleService.kt
```diff
@@ -1,72 +0,0 @@
...
-import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.AssessmentStatus.OPTED_OUT
-import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.StatusChangedEvent
-import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.model.EligibilityStatus.ELIGIBLE
-import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.model.EligibilityStatus.INELIGIBLE
-import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.model.EligibilityStatus.IN_PROGRESS
-import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.service.AssessmentService.AssessmentWithEligibilityProgress
-import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.service.StatusHelpers.calculateAggregateEligibilityStatus
-
-@Service
-class AssessmentLifecycleService {
```
andrewrlee, line 22: no longer need all this weird stuff

PR #65, opened 2024-11-26
Cvsl 2263 afer com residential checks listen to events to update com on allocation events

nishanthmahasamudram:


justin-gardiner (approved):
LGTM

PR #71, opened 2024-12-05
Cvsl 2261 endpoint to get task config

justin-gardiner:


detekt-baseline.xml
```diff
@@ -22,5 +22,6 @@
     <ID>SpreadOperator:HmppsAssessForEarlyReleaseApi.kt$(*args)</ID>
     <ID>TooManyFunctions:AddressService.kt$AddressService</ID>
     <ID>UnusedParameter:ResidentialChecksService.kt$ResidentialChecksService$addressCheckRequestId: Long</ID>
+    <ID>UnusedParameter:ResidentialChecksService.kt$ResidentialChecksService$requestId: Long</ID>
```
justin-gardiner, line 25: This is just temporary as when we are retrieving task status from the db etc this parameter will be used.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/policy/ResidentialChecksPolicyV1.kt
```diff
@@ -3,22 +3,22 @@ package uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.service.polic
...
   tasks = listOf(
-    ResidentialChecksTask(
+    Task(
       code = "address-details-and-informed-consent",
       name = "Address details and informed consent",
       sections = listOf(
         Section(
           questions = listOf(
-            Question(
+            TaskQuestion(
```
justin-gardiner, line 21: Had to rename this as there was already a Question class for the eligibility and suitability checks and this overwriting that one in the type def file for the front end. I don't know if there's a way around this?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/EligibilityAndSuitabilityCaseViewResourceIntTest.kt
```diff
@@ -79,7 +79,7 @@ class EligibilityAndSuitabilityCaseViewResourceIntTest : SqsIntegrationTestBase(
         .exchange()
         .expectStatus()
         .isOk
-        .expectBody().json(serializedContent("eligibility-and-suitability-view"), true)
```
justin-gardiner, line 82: the overload that takes a boolean is deprecated

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/ResidentialChecksTaskView.kt
```diff
@@ -0,0 +1,19 @@
...
+)
+data class ResidentialChecksTaskView(
+  @Schema(description = "A summary of an offender's current assessment")
+  val assessmentSummary: AssessmentSummary,
+
+  @Schema(description = "Details of the task")
+  val taskConfig: Task,
+
+  @Schema(description = "The current status of the task")
+  val taskStatus: TaskStatus,
```
justin-gardiner, line 18: We will need to add answers etc later

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/policy/ResidentialChecksPolicyV1.kt
andrewrlee: oh yes - this is a really long outstanding issue with openapi, I think it might be possible to change it to use a fully qualified name but then it'd be quite difficult to read

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/EligibilityAndSuitabilityCaseViewResourceIntTest.kt
andrewrlee: does it default to strict checking?

justin-gardiner: nope it doesn't and now I've just looked it's not deprecated in the version we're using. I must have done this when I was looking at upgrading dps spring boot plugin. I've reverted.

andrewrlee (approved):
LGTM

PR #75, opened 2024-12-10
query update

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/OffenderService.kt
```diff
@@ -59,6 +59,31 @@ class OffenderService(
...
+  fun getDecisionMakerCaseload(prisonCode: String): List<OffenderSummary> {
+    val decisionMakerStatuses = listOf(
+      AssessmentStatus.AWAITING_DECISION,
+      AssessmentStatus.AWAITING_REFUSAL,
+      AssessmentStatus.APPROVED,
+      AssessmentStatus.REFUSED,
+      AssessmentStatus.OPTED_OUT,
+      AssessmentStatus.TIMED_OUT,
+      AssessmentStatus.POSTPONED,
+    )
```
andrewrlee, line 72: This is good for now but think we could maybe move this to be an attribute on status: `val visibleToDm: Boolean` and maybe have similar for PP and CA or a `visibleToRole: map<Role, Boolean>` 
(though it might be that CA can see every status anyway)

andrewrlee (approved):
LGTM

PR #76, opened 2024-12-11
Cvsl 2365 afer dm add caseload endpoint

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/repository/OffenderRepository.kt
```diff
@@ -2,12 +2,12 @@ package uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.repository
...
 import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.AssessmentStatus
 import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.Offender
-import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.OffenderStatus
 
 @Repository
 interface OffenderRepository : JpaRepository<Offender, Long> {
   fun findByPrisonNumber(prisonerNumber: String): Offender?
 
-  fun findByPrisonIdAndStatus(prisonId: String, status: OffenderStatus): List<Offender>
```
nishanthmahasamudram, line 12: replaced this to use single point of source.
Is this ok?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AssessmentStatus.kt
```diff
@@ -89,21 +110,68 @@ enum class AssessmentStatus {
...
+  AWAITING_PRE_RELEASE_CHECKS {
+    override val visibleToRole = mapOf(
+      PRISON_CA to true,
+      PROBATION_COM to false,
+      PRISON_DM to false,
+    )
+  },
 
-  PASSED_PRE_RELEASE_CHECKS,
+  PASSED_PRE_RELEASE_CHECKS {
```
andrewrlee, line 152: I know this was my orig suggestion but I think this would probably be slightly more elegant as a set of roles that it's visible to

But this is fine for now! 
Looks good!

andrewrlee (approved):
LGTM

PR #81, opened 2024-12-20
Update dependencies

andrewrlee:


build.gradle.kts
```diff
@@ -42,13 +42,15 @@ detekt {
   baseline = file("$projectDir/detekt-baseline.xml") // a way of suppressing issues before introducing detekt
 }
 
+ext["hibernate.version"] = "6.5.3.Final"
```
andrewrlee, line 45: pinning hibernate as the latest version doesn't handle manually setting ids without use of a [custom sequence generator](https://hibernate.atlassian.net/browse/HHH-17472). Support for this will be included by default in hibernate 7

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/event/PrisonOffenderEventListener.kt
```diff
@@ -32,7 +30,6 @@ class PrisonOffenderEventListener(
   }
 
   @SqsListener("domaineventsqueue", factory = "hmppsQueueContainerFactoryProxy")
-  @WithSpan(value = "hmpps-assess-for-early-release-prisoner-domain-event-queue", kind = SpanKind.SERVER)
```
andrewrlee, line 35: these spans are auto registered by the sqs library now so need to remove these to prevent double logging

applicationinsights.dev.json
```diff
@@ -42,6 +42,17 @@
...
+      {
+        "telemetryType": "exception",
+        "attributes": [
+          {
+            "key": "exception.type",
+            "value": "java.util.concurrent.CompletionException",
+            "matchType": "strict"
+          }
+        ],
+        "percentage": 10
```
andrewrlee, line 55: need to drastically reduce these as these will be logged twice for the majority of cases. 
There's a minor edge case where these wouldn't be logged at all without which is the reason we're not reducing to 0%

nishanthmahasamudram (approved):
LGTM

PR #83, opened 2024-12-30
Cvsl 2211 afer set up gotenberg for pdf generation

nishanthmahasamudram:


build.gradle.kts
```diff
@@ -64,6 +64,13 @@ dependencies {
   // OpenAPI
   implementation("org.springdoc:springdoc-openapi-starter-webmvc-ui:2.7.0")
 
+  // OkHttp
+  implementation("com.squareup.okhttp3:okhttp:4.9.3")
+  implementation("com.squareup.okio:okio:2.10.0")
```
andrewrlee, line 69: cool, so we'll get rid of this in a follow up and move to spring?

docker-compose.yml
```diff
@@ -44,6 +44,21 @@ services:
       timeout: 5s
       retries: 5
 
+
+
+  gotenberg:
```
andrewrlee, line 49: meh.. couple of extra lines here.. if you're doing a follow up could tidy this up?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/resource/PdfResource.kt
```diff
@@ -0,0 +1,77 @@
...
+        content = [
+          Content(
+            mediaType = "application/json",
+            schema = Schema(implementation = ErrorResponse::class),
+          ),
+        ],
+      ),
+    ],
+  )
+  fun getPdf(@RequestParam title: String, @RequestParam message: String): ByteArray? {
```
andrewrlee, line 74: when we implement a form then we'd probably need to pass through a form name/id/code and the assessment id.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/resource/PdfResource.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.config.ErrorResponse
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.service.PdfService
+
+@RestController
+@RequestMapping(produces = [MediaType.APPLICATION_JSON_VALUE])
+class PdfResource(
```
andrewrlee, line 20: `FormResource` may be a better name?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/PdfService.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    return try {
+      val response: ResponseEntity<ByteArray> = restTemplate.postForEntity(
+        "$gotenbergHost/forms/chromium/convert/html",
+        requestEntity,
+        ByteArray::class.java,
+      )
+      response.body
+    } catch (e: HttpClientErrorException) {
+      log.error("Error converting HTML to PDF", e)
```
andrewrlee, line 69: probably better to throw this (not handle it) - we'd rather a 500 response be generated by the exception handler

build.gradle.kts
andrewrlee: ah.. so this isn't being used anymore.. so safe to remove

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/PdfService.kt
```diff
@@ -0,0 +1,73 @@
...
+    body.add("paperHeight", "11.69")
+    body.add("marginTop", "1")
+    body.add("marginBottom", "1")
+    body.add("marginLeft", "1")
+    body.add("marginRight", "1")
+
+    val requestEntity = HttpEntity(body, headers)
+
+    return try {
+      val response: ResponseEntity<ByteArray> = restTemplate.postForEntity(
```
andrewrlee, line 62: restTemplate is a bit of an older way of doing things, it was going to be deprecate but is now in [maintenance mode](https://stackoverflow.com/questions/47974757/webclient-vs-resttemplate). We should probably use web client to be consistent with the other apis but this isn't of huge importance now

andrewrlee (dismissed):
LGTM

helm_deploy/values-dev.yaml
```diff
@@ -16,6 +16,15 @@ generic-service:
     HMPPS_PRISONREGISTER_API_URL: "https://prison-register-dev.hmpps.service.justice.gov.uk"
     HMPPS_PROBATIONSEARCH_API_URL: "https://probation-offender-search-dev.hmpps.service.justice.gov.uk"
     OS_PLACES_API_URL: "https://api.os.uk/search/places/v1"
+    GOTENBERG_API_URL: "http://hmpps-assess-for-early-release-gotenberg"
+
+gotenberg:
```
andrewrlee, line 21: do you need to specify an image tag in values.yml?

docker-compose-test.yml
```diff
@@ -20,6 +20,19 @@ services:
     volumes:
       - test-pgdata:/var/lib/postgresql/data
 
+  gotenberg:
+    image: gotenberg/gotenberg:7
```
andrewrlee, line 24: looks like there's a [gotenberg 8](https://github.com/gotenberg/gotenberg/releases/tag/v8.0.0) now, could address in follow up?

andrewrlee (approved):
LGTM

PR #92, opened 2025-01-13
CVSL-2391: Add custom validation messages to residential checks task …

justin-gardiner:
…answers. Customise open API type generated for Map<String, Any>

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/residentialChecks/AddressDetailsTaskAnswers.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/residentialChecks/AddressDetailsTaskAnswers.kt
@@ -7,7 +7,7 @@ import jakarta.persistence.Entity
 import jakarta.validation.constraints.NotNull
 import org.hibernate.annotations.Type
 import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.CurfewAddressCheckRequest
-import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.service.policy.VisitedAddress
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.service.policy.model.residentialchecks.VisitedAddress
 
 @Entity
 @DiscriminatorValue(value = "address-details-and-informed-consent")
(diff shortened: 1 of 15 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/residentialChecks/AddressDetailsTaskAnswers.kt
```diff
@@ -33,14 +33,14 @@ class AddressDetailsTaskAnswers(
 }
 
 data class AddressDetailsAnswers(
-  @field:NotNull
-  val electricitySupply: Boolean?,
+  @field:NotNull(message = "Select if the address is connected to an electricity supply")
+  val electricitySupply: String?,
```
justin-gardiner, line 37: Changed to String so we store the actual value selected by the user like "Yes" instead of storing at True and having to convert back and forth.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/residentialChecks/AddressDetailsTaskAnswers.kt
```diff
@@ -33,14 +33,14 @@ class AddressDetailsTaskAnswers(
...
 
-  @field:NotNull
-  val visitedAddress: Boolean?,
+  @field:NotNull(message = "Select if you have visited the address")
```
justin-gardiner, line 39: These are the validation messages displayed in the front end if backend validation fails.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/residentialChecks/AddressDetailsTaskAnswers.kt
andrewrlee: Would be better if we could store booleans.. Could we use differentiate by having different `value` and `name`?
I'm sure it might be a bit more difficult than just doing that

justin-gardiner: ok, I could probably add a Jacskson deserializer so that the front end sends what is input on the form, Yes / No and Jackson converts this to a boolean.

justin-gardiner: Actually, I might have another solution. I will spend a bit of time trying it

justin-gardiner: Done, back to booleans :)

andrewrlee (dismissed):
LGTM

nishanthmahasamudram (approved):
LGTM

PR #93, opened 2025-01-13
CVSL-2418 add database_schema_report orb

gurps-singh:
This PR is to add the database schema report orb to the AfER pipeline.


.circleci/config.yml
```diff
@@ -93,6 +93,13 @@ workflows:
               ignore: /.*/
       - hmpps/helm_lint:
           name: helm_lint
+      - hmpps/database_schema_report:
+          schema: "public"
+          initialise_database: "./gradlew check"
```
andrewrlee, line 98: this will rerun all the tests again, ktlint, detekt etc, which it would be good to avoid.. think this is why they just chose to implement the single test / gradle task

PR #99, opened 2025-01-16
CVSL-2418 only run report job on main branch

gurps-singh:
This PR is to ensure that the database schema job is only run when a main run is happening on Circle.


andrewrlee (approved):
LGTM

PR #102, opened 2025-01-23
Cvsl 2391 address check status updates

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentService.kt
```diff
@@ -97,6 +100,29 @@ class AssessmentService(
...
+  @Transactional
+  fun submitForPreDecisionChecks(prisonNumber: String) {
+    val assessmentEntity = getCurrentAssessment(prisonNumber).assessmentEntity
+    assessmentEntity.performTransition(CompleteAddressChecks)
+    assessmentRepository.save(assessmentEntity)
+  }
+
+  @Transactional
+  fun updateAddressChecksStatus(prisonNumber: String, status: ResidentialChecksStatus) {
+    val currentAssessment = getCurrentAssessment(prisonNumber)
```
andrewrlee, line 112: does this mean that the frontend needs to work out whether the checks are complete or not? 
how does it do that?

andrewrlee: ah no it's in the ResidentialChecksService

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentService.kt
```diff
@@ -97,6 +100,29 @@ class AssessmentService(
...
+    currentAssessment.assessmentEntity.performTransition(
+      AssessmentLifecycleEvent.ResidentialCheckStatusAnswerProvided(status),
+    )
+
+    val assessmentEntity = currentAssessment.assessmentEntity
+    if (status == ResidentialChecksStatus.SUITABLE && !assessmentEntity.addressChecksComplete) {
```
andrewrlee, line 118: mmh. you could possibly derive this from the result of the perform transition.. that would say whether a transition has occured or not? 
I'm not sure if its "better" to do it that way or not!

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentService.kt
andrewrlee: maybe it removes the need to do the checks twice?

justin-gardiner: I'll have to think about that. We don't do a state transition when all tasks are complete. We still stay in ADDRESS_AND_RISK_CHECKS_IN_PROGRESS. So I can't see how we can look at a state transition, and therefore  we need this extra flag to say all checks have been completed and all are suitable so enable the user to send to prison.

PR #103, opened 2025-01-28
CVSL-2395-AFER-Add-other-occupants-for-CAs-when-they-add-an-address-c…

nishanthmahasamudram:
…heck


src/test/resources/test_data/reset.sql
```diff
@@ -1,5 +1,6 @@
 delete from assessment_event;
 delete from resident;
+TRUNCATE TABLE resident RESTART IDENTITY CASCADE;
```
nishanthmahasamudram, line 3: If not adding this, I am facing intermittent issues with resident ID's. 
They are not resetting even after delete command.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/prison/AddressService.kt
```diff
@@ -132,22 +132,49 @@ class AddressService(
...
+
+    // Delete residents not present in addResidentsRequest
+    if (residentsToDelete.isNotEmpty()) {
+      residentRepository.deleteAll(residentsToDelete)
     }
 
-    return residents
+    val residentsToSave = addResidentsRequest.map { addResidentRequest ->
+      if (recordsToUpdate.any { it.id == addResidentRequest.residentId }) {
+        recordsToUpdate.find { it.id == addResidentRequest.residentId }?.apply {
```
andrewrlee, line 151: rather than loop twice could `find` and then null check

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/AddressResourceIntTest.kt
```diff
@@ -434,7 +434,7 @@ class AddressResourceIntTest : SqsIntegrationTestBase() {
       val standardAddressCheckRequestSummary = checkRequest as StandardAddressCheckRequestSummary
       assertThat(standardAddressCheckRequestSummary.address.firstLine).isEqualTo("4 ADANAC DRIVE")
       val residentSummary = standardAddressCheckRequestSummary.residents
-      assertThat(residentSummary).hasSize(2)
+      assertThat(residentSummary).hasSize(3)
```
andrewrlee, line 437: It'd probably be worth having a test where you delete a resident too?

andrewrlee: though guess that would be in the [AddressServiceIntTest.kt](https://github.com/ministryofjustice/hmpps-assess-for-early-release-api/pull/103/files#diff-f988b7f5f8e59cad9ddd700131f09bd940792d20a863825a02f09ff42f8c9c16)

src/test/resources/test_data/reset.sql
andrewrlee: ah yes this is a common issue - the sequences don't get reset.

andrewrlee: Might be nicer to do a `ALTER SEQUENCE seq RESTART WITH 1;` where `seq` is the name of the sequence.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/AddressResourceIntTest.kt
nishanthmahasamudram: Yes it is in [AddressServiceIntTest.kt](https://github.com/ministryofjustice/hmpps-assess-for-early-release-api/pull/103/files#diff-f988b7f5f8e59cad9ddd700131f09bd940792d20a863825a02f09ff42f8c9c16)

andrewrlee (approved):
LGTM

PR #110, opened 2025-02-11
CVSL-2392: Add working days to HDCED to OffenderSummary.

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/workingdays/BankHolidayService.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.springframework.cache.annotation.Cacheable
+import org.springframework.stereotype.Service
+
+@Service
+class BankHolidayService(
+  private val govUkApiClient: GovUkApiClient,
+) {
+
+  @Cacheable("bank-holidays")
+  fun getBankHolidaysForEnglandAndWales() = govUkApiClient.getBankHolidaysForEnglandAndWales()
```
andrewrlee, line 12: 🥳

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/workingdays/WorkingDaysServiceTest.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    @Test
+    fun `No working days before today`() {
+      val workingDaysBefore = service.workingDaysBefore(today)
+      assertThat(workingDaysBefore).isZero()
+    }
+
+    @Test
+    fun `bank holidays do not count as working days`() {
+      org.mockito.kotlin.reset(bankHolidayService)
```
andrewrlee, line 84: is this needed? doesn't the before each work?

andrewrlee (dismissed):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/workingdays/WorkingDaysServiceTest.kt
justin-gardiner: we also set up a mock response in the before each after the reset, so without this, it would return that response for the first call followed by this for the second. I want a different set of bank holidays for this test

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/OffenderService.kt
```diff
@@ -154,6 +156,7 @@ class OffenderService(
     forename = offender.forename,
     surname = offender.surname,
     hdced = offender.hdced,
+    workingDaysToHdced = workingDaysService.workingDaysBefore(offender.hdced),
```
andrewrlee, line 159: guess this isn't covered in the current tests?

justin-gardiner: Nope, we were only checking booking id was correct, updated to check OffenderSummary is correctly populated.

andrewrlee (approved):
LGTM

PR #112, opened 2025-02-12
Defect/cvsl 2519 up date to assessment enitiy to include opt out details

AledWEvans:
CVSL-2519 AFER | Capture Opt out reason when a case administrator opts out a case

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentService.kt
@@ -103,13 +103,10 @@ class AssessmentService(
 
   @Transactional
   fun optOut(prisonNumber: String, optOutRequest: OptOutRequest) {
-    transitionAssessment(prisonNumber, OptOut, optOutRequest.agent)
-
     val assessmentEntity = getCurrentAssessment(prisonNumber).assessmentEntity
-
+    assessmentEntity.performTransition(OptOut, optOutRequest.agent.toEntity())
     assessmentEntity.optOutReasonType = optOutRequest.reasonType
     assessmentEntity.optOutReasonOther = optOutRequest.otherDescription
-
     assessmentRepository.save(assessmentEntity)
   }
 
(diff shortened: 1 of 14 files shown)


src/test/resources/test_data/responses/eligibility-and-suitability-view.json
```diff
@@ -11,6 +11,8 @@
     "responsibleCom": null,
     "team": null,
     "policyVersion": "1.0",
+    "optOutReasonType":null,
```
AledWEvans, line 14: Will add decent values here, just want you guys to have a look, to tell me if in general I'm on the right track for this change?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentService.kt
```diff
@@ -102,6 +104,13 @@ class AssessmentService(
   @Transactional
   fun optOut(prisonNumber: String, optOutRequest: OptOutRequest) {
     transitionAssessment(prisonNumber, OptOut, optOutRequest.agent)
+
+    val assessmentEntity = getCurrentAssessment(prisonNumber).assessmentEntity
```
andrewrlee, line 108: I think it'd be better to extract out the transition assessment.. to avoid the two db look ups and saves.

```
... getcurrentAssessment
transition
set opt out reason
set opt out reasonn other
save..
```

andrewrlee: I get this is basically doing the same as from `updateAddressChecksStatus`

andrewrlee (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentService.kt
AledWEvans: Yes did copy that code have updated to your suggested approach, looks better makes sense

andrewrlee (approved):
LGTM

PR #117, opened 2025-02-20
CVSL-2496: Add current task to OffenderSummary.

justin-gardiner:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AssessmentTest.kt
```diff
@@ -192,4 +212,28 @@ class AssessmentTest {
     val statusChangeEvents = assessment.assessmentEvents.map { it as StatusChangedEvent }.map { it.changes }
     assertThat(statusChangeEvents).isEmpty()
   }
+
+  @Test
```
justin-gardiner, line 216: Some of these are wrong as we're missing tasks etc.

andrewrlee (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AssessmentTest.kt
andrewrlee: I think it blows up loudly to highlight it's not mapped

andrewrlee: Guess this might be a bit more elegant expressed as a 
```
 @ParameterizedTest
    @CsvSource(
```

justin-gardiner: Yes good point, will update when we add the new tasks etc.

PR #119, opened 2025-02-20
CVSL-2549 AFER | Add missing profile details to CA Case View

AledWEvans:
Some refactoring as the mapping to the summary object was in two different places.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/mapper/OffenderToAssessmentSummaryMapper.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+    val prisonerDetails = prisonerSearchResults.first()
+    return AssessmentSummary(
+      forename = offender.forename,
+      surname = offender.surname,
+      dateOfBirth = offender.dateOfBirth,
+      prisonNumber = offender.prisonNumber,
+      hdced = offender.hdced,
+      crd = offender.crd,
+      location = prisonerDetails.prisonName,
```
AledWEvans, line 30: Making assumptions that the prison name will be correct from the prisoner details query.
Could be problems when the prisoner is being temporary moved, but this should not be a problem at this stage in the sentence.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/mapper/OffenderToAssessmentSummaryMapper.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.Offender
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.exception.PrisonerNotFoundException
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.model.AssessmentSummary
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.model.TaskProgress
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.model.toSummary
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.service.prison.PrisonService
+
+@Component
+class OffenderToAssessmentSummaryMapper(
```
AledWEvans, line 12: Is this a builder or a mapper or both :-/

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentService.kt
```diff
@@ -24,71 +24,52 @@ import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.model.OptOutRe
...
+    ?: throw EntityNotFoundException("Cannot find offender with prisonNumber $prisonNumber")
+
+  @Transactional
+  fun getCurrentAssessment(prisonNumber: String) = getOffender(prisonNumber).currentAssessment()
+
   @Transactional
-  fun getCurrentAssessment(prisonNumber: String): AssessmentWithEligibilityProgress {
-    val offender = offenderRepository.findByPrisonNumber(prisonNumber)
-      ?: throw EntityNotFoundException("Cannot find offender with prisonNumber $prisonNumber")
+  fun getCurrentAssessmentWithEligibilityProgress(prisonNumber: String): AssessmentWithEligibilityProgress {
```
AledWEvans, line 58: I wonder if this should be moved into the EligibilityAndSuitabilityService?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/mapper/OffenderToAssessmentSummaryMapper.kt
```diff
@@ -0,0 +1,44 @@
...
+  private val prisonService: PrisonService,
+) {
+
+  fun map(offender: Offender): AssessmentSummary {
+    val currentAssessment = offender.currentAssessment()
+    val prisonerSearchResults = prisonService.searchPrisonersByNomisIds(listOf(offender.prisonNumber))
```
andrewrlee, line 18: It'd be good to bulk call this rather than one per prisoner

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/config/HmppsAssessForEarlyReleaseApiExceptionHandler.kt
```diff
@@ -52,6 +53,17 @@ class HmppsAssessForEarlyReleaseApiExceptionHandler {
       ),
     ).also { log.info("No resource found exception: {}", e.message) }
 
+  @ExceptionHandler(PrisonerNotFoundException::class)
+  fun handlePrisonerNotFoundException(e: PrisonerNotFoundException): ResponseEntity<ErrorResponse> = ResponseEntity
```
andrewrlee, line 57: we use `EntityNotFoundException` in a few places, would probably be better to not have to handle different types of missing resources -so could have a new generic exception or use EntityNotFoundException

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/mapper/OffenderToAssessmentSummaryMapper.kt
andrewrlee: or is this currently _only_ used for getting a single prisoners details?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/config/HmppsAssessForEarlyReleaseApiExceptionHandler.kt
AledWEvans: Did wonder, I just struggled over weather it was an Entity, So how about ItemNotFoundException?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/mapper/OffenderToAssessmentSummaryMapper.kt
andrewrlee: Could we grab this as previously so it's consistent with the current assessment view?
 `val prisonName = prisonService.getPrisonNameForId(offender.prisonId)`

(we handle transfers separately)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/mapper/OffenderToAssessmentSummaryMapper.kt
AledWEvans: I think it's only for one!?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/mapper/OffenderToAssessmentSummaryMapper.kt
AledWEvans: Yes no probs

andrewrlee (approved):
LGTM

PR #122, opened 2025-02-24
CVSL-2549 AFER | cell location and main offense are optional

AledWEvans:
Have made these two optional 
```
  val cellLocation: String? = null,
  val mostSeriousOffence: String? = null,
```
as they are not always there during transfer.. Also wonder if mostSeriousOffence is correct element to use :
```
 "allConvictedOffences": [
      {
        "primarySentence": true
      }
    ]
```
This fix is to allow the server side code to work while I investigate

Should have spotted this will install postman in future


andrewrlee (approved):
LGTM

PR #124, opened 2025-02-27
CVSL-2500: Record agent when saving eligibility and suitability answers.

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/StatusChangedEvent.kt
```diff
@@ -17,7 +17,7 @@ class StatusChangedEvent(
   @Column(columnDefinition = "jsonb")
   val changes: StatusChange,
 
-  agent: Agent = Agent(UserRole.SYSTEM.name, UserRole.SYSTEM, UserRole.SYSTEM.name),
+  agent: Agent = Agent(UserRole.SYSTEM.name, UserRole.SYSTEM.name, UserRole.SYSTEM, UserRole.SYSTEM.name),
```
andrewrlee, line 20: guess we could use SYSTEM_AGENT constant..

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/StatusChangedEvent.kt
justin-gardiner: Yes, I'll add a TODO locally, so I do now that on my next PR

PR #126, opened 2025-02-27
CVSL-2500: Return last updated date in eligibility and suitability ca…

justin-gardiner:
…se view.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/EligibilityAndSuitabilityCaseView.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/EligibilityAndSuitabilityCaseView.kt
@@ -1,8 +1,10 @@
 package uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.model
 
+import com.fasterxml.jackson.annotation.JsonFormat
 import com.fasterxml.jackson.annotation.JsonProperty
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/EligibilityAndSuitabilityCaseView.kt
```diff
@@ -88,6 +90,9 @@ data class EligibilityCriterionProgress(
   val questions: List<Question> = emptyList(),
   @Schema(description = "Details of the user that submitted the answers for this criterion")
   val agent: Agent?,
+  @Schema(description = "The date time when answers were last submitted for this criterion or null if no answers have been submitted yet", example = "16/08/2025")
+  @JsonFormat(pattern = "yyyy-MM-dd")
+  val lastUpdated: LocalDate?,
```
andrewrlee, line 95: guess technically it's a date rather than date time...

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/EligibilityAndSuitabilityCaseView.kt
justin-gardiner: Ahh, good spot, thought I had fixed this, will do on the next PR

PR #127, opened 2025-02-28
CVSL-2543 AFER | Capture CRN

AledWEvans:
We need to record CRN when creating an offender record 

we already retrieve it but throw away, just need to refactor 

1. Need to persist on Offender as crn
2. Expose via OffenderSummary representation
3. Handle missing connections between nomis and delius data - make CRN nullable to handle these cases. 

**I have maid caseReferenceNumber nullable for missing nomis data, please check if that will cover option 3.**

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Offender.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Offender.kt
@@ -43,7 +43,7 @@ data class Offender(
 
   val crd: LocalDate? = null,
 
-  val caseReferenceNumber: String? = null,
+  val crn: String? = null,
 
   val sentenceStartDate: LocalDate? = null,
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Offender.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Offender.kt
@@ -43,6 +43,8 @@ data class Offender(
 
   val crd: LocalDate? = null,
 
+  val caseReferenceNumber: String? = null,
+
   val sentenceStartDate: LocalDate? = null,
 
   @OneToMany(mappedBy = "offender", fetch = FetchType.LAZY, cascade = [CascadeType.ALL], orphanRemoval = true)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/OffenderService.kt
@@ -39,19 +39,19 @@ class OffenderService(
   @Transactional
   fun getCaseAdminCaseload(prisonCode: String): List<OffenderSummaryResponse> {
     val assessments = assessmentRepository.findByOffenderPrisonId(prisonCode)
-    return assessments.map { createOffenderSummaryResponse(it) }
+    return assessments.map { it.toOffenderSummary() }
   }
 
   @Transactional
   fun getComCaseload(staffCode: String): List<OffenderSummaryResponse> {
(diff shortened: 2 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/OffenderService.kt
```diff
@@ -154,21 +162,23 @@ class OffenderService(
...
-    workingDaysToHdced = workingDaysService.workingDaysBefore(offender.hdced),
-    probationPractitioner = this.responsibleCom?.fullName,
-    isPostponed = this.status == AssessmentStatus.POSTPONED,
-    postponementDate = this.postponementDate,
-    postponementReasons = this.postponementReasons.map { reason -> reason.reasonType }.toList(),
-    status = this.status,
-    addressChecksComplete = this.addressChecksComplete,
-    currentTask = this.currentTask(),
-    taskOverdueOn = offender.sentenceStartDate?.plusDays(DAYS_BEFORE_SENTENCE_START),
+  private fun createOffenderSummaryResponse(assessment: Assessment): OffenderSummaryResponse = OffenderSummaryResponse(
```
andrewrlee, line 165: was there a reason you changed this from an extension function? 
We use this pattern all over the place

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Offender.kt
```diff
@@ -43,6 +43,8 @@ data class Offender(
 
   val crd: LocalDate? = null,
 
+  val caseReferenceNumber: String? = null,
```
andrewrlee, line 46: style guides suggest this should generally be referred to as `crn`. 
It's how probation managers refer to it

andrewrlee: https://design-patterns.service.justice.gov.uk/content-standards/style-guide/

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/OffenderService.kt
AledWEvans: Generally the convention i think is the the service or a mapper/builder should know how to convert something not one of  the objects used to create the offender summary. Extension function are powerful but can be difficult to maintain and can break rules of responsibility.  Also if we used a mapper/builder approach in separate class we can test independently.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Offender.kt
AledWEvans: Sure will update.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/OffenderService.kt
andrewrlee: It's scoped to the service so the responsibility and logic is sat within the service rather than domain entity and the domain entity still doesn't know anything about the api or response objects

andrewrlee (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/OffenderService.kt
AledWEvans: I get the scope thing,... but why not just add it to a mapper, were every one knows were it should be implemented what does it actually give us in this case and what if some one else add's another extension with same name in a different scope... my be I just don't like extension function unless they are used in a very particular way. Yet's have a chat before we merge!?

andrewrlee (approved):
LGTM

PR #128, opened 2025-03-03
Update to docs and a script to create an .env file

AledWEvans:
Just added a script to quickly set up running spring application from with in the IDE and improved docs on the readme file.


andrewrlee (approved):
LGTM

PR #133, opened 2025-03-12
CVSL-2607-AFER-CA-view-task-list-for-eligibility-and-suitability-prog…

nishanthmahasamudram:
…ress


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentService.kt
```diff
@@ -191,4 +213,21 @@ class AssessmentService(
       lastUpdated = eligibilityCheckResult?.lastUpdatedTimestamp?.toLocalDate(),
     )
   }
+
+  private fun getPrisonerDetails(offender: Offender): List<PrisonerSearchPrisoner> {
```
andrewrlee, line 217: could return first here rather than a list

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AssessmentService.kt
nishanthmahasamudram: will fix it in my next PR.

PR #141, opened 2025-03-18
CVSL-2700 Moving event handling out of state machine

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/state/AssessmentLifecycleEvent.kt
```diff
@@ -1,19 +1,33 @@
...
+
+  data class EligibilityChecksPassed(
+    val type: CriteriaType,
+    val code: String,
+    val answers: Map<String, Boolean>,
+  ) : AssessmentLifecycleEvent() {
+    override fun getContext(): Map<String, Any> = mapOf("type" to type, "code" to code, "answers" to answers)
+  }
+
+  data class EligibilityChecksFailed(
```
justin-gardiner, line 25: This looks exactly the same as EligibilityChecksPassed above, unless I'm missing something?

justin-gardiner (approved):
LGTM

PR #142, opened 2025-03-18
CVSL-2700 remove address state changes from machine

andrewrlee:
We'll need to handle the transition when all addresses are unsuitable


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/state/AssessmentStateMachine.kt
```diff
@@ -66,27 +65,13 @@ val assessmentStateMachine = StateMachine.create<AssessmentState, AssessmentLife
...
   state<AwaitingAddressAndRiskChecks> {
-    on<ResidentialCheckStatusAnswerProvided> {
-      when (it.checkStatus) {
-        ResidentialChecksStatus.UNSUITABLE -> transitionTo(AddressAndRiskChecksInProgress)
-        ResidentialChecksStatus.IN_PROGRESS -> transitionTo(AddressAndRiskChecksInProgress)
-        ResidentialChecksStatus.SUITABLE -> dontTransition(Error("Unable to transition to suitable from ${this.status} directly"))
-        else -> error("Unexpected eligibility status: $it")
-      }
-    }
+    on<ResidentialCheckAnswerProvided> { transitionTo(AddressAndRiskChecksInProgress) }
```
andrewrlee, line 68: We just need to track whether they've started at the moment and this will allow the transition to handover back to CA. 

We'll need to also handle when all address checks are unsuitable later on.

AledWEvans (approved):
LGTM

PR #144, opened 2025-03-20
CVSL-2672 AfER | Support soft delete of assessments, This is a draft …

AledWEvans:
This is a possible way of tranfering address and check over to the next assessment,.. The code has not been run because I thought it would be better to make sure I'm on the right track before I commit more time to this approach

### DIFF
--- a/src/main/resources/migration/common/v49__add_address_to_assessments_table.sql
+++ b/src/main/resources/migration/common/v49__add_address_to_assessments_table.sql
@@ -5,6 +5,9 @@ CREATE TABLE address_to_assessments (
         CONSTRAINT address_to_assessments_pkey PRIMARY KEY (check_request_id, assessment_id)
 );
 
+-- rename table to standard naming used for derived tables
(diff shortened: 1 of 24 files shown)


src/main/resources/migration/common/v49__add_address_to_assessments_table.sql
```diff
@@ -0,0 +1,18 @@
...
+        check_request_id int8 NOT NULL,
+        assessment_id int8 NOT NULL,
+        CONSTRAINT address_to_assessments_pkey PRIMARY KEY (check_request_id, assessment_id)
+);
+
+-- rename table to standard naming used for derived tables
+ALTER TABLE cas_check_request RENAME TO cas_address_check_request;
+
+-- Migrate data across
+INSERT INTO address_to_assessments (check_request_id, assessment_id)
```
AledWEvans, line 12: This should be an update

AledWEvans:
This is not needed, as the point on the ticket was to prevent the cascade of the soft delete to the child entities.

PR #146, opened 2025-03-21
Cvsl 2674 afer tech move booking id to assessment

AledWEvans:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/AssessmentSummary.kt
```diff
@@ -8,6 +8,10 @@ import java.time.LocalDate
 
 @Schema(description = "Response object which describes an assessment")
 data class AssessmentSummary(
```
nishanthmahasamudram, line 10: can you add the same to AssessmentOverviewSummary ?

andrewrlee (approved):
LGTM

PR #148, opened 2025-03-21
CVSL-2672 AfER | Support soft delete of assessments - some how when I…

AledWEvans:
When I merged main into my PR I removed the test, now i have added it back in


andrewrlee (approved):
LGTM

PR #150, opened 2025-03-24
CVSL-2684: Add information about address check form.

justin-gardiner:


src/test/resources/test_data/thymeleaf_html/offender_address_checks_information_form.html
```diff
@@ -5,20 +5,55 @@
...
+    <div>Location: Birmingham (HMP)</div>
+    <div>Prison no: A1234AA</div>
+    <div>Date: ((current_date))</div>
   </div>
 </div>
-<div>
-  <div>
-    <div>20-03-1978</div>
-  </div>
+<p>24 Mar 2025</p>
```
AledWEvans, line 17: This should be as below... as dateToday is based on todays date
`<p>((current_date))</p>`

AledWEvans (changes requested):
Some changes needed, A test will fail tomorrow using the current code

src/test/resources/test_data/thymeleaf_html/offender_address_checks_information_form.html
justin-gardiner: Good spot, fixed

AledWEvans (approved):
LGTM

PR #152, opened 2025-03-26
CVSL-2667: Add crd to OffenderSummaryResponse.

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/OffenderSummaryResponse.kt
```diff
@@ -21,6 +21,10 @@ data class OffenderSummaryResponse(
   @Schema(description = "The offender's surname", example = "Smith")
   val surname: String,
 
+  @Schema(description = "The offender's conditional release date date", example = "2026-08-23")
+  @JsonFormat(pattern = "yyyy-MM-dd")
+  val crd: LocalDate? = null,
```
AledWEvans, line 26: I would have thought there was a test update for this value

AledWEvans (changes requested):
minor change might be needed

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/OffenderSummaryResponse.kt
justin-gardiner: Added a test

AledWEvans (approved):
LGTM

PR #167, opened 2025-04-07
CVSL-2722 remove status from offender table and entity

gurps-singh:
This PR is to remove the status column on the Offender entity as it belongs on the assessment (and is already present there).


AledWEvans (approved):
LGTM

PR #175, opened 2025-04-22
CVSL-2776-AfER-Update-reasons-for-ineligibility-unsuitability-on-the-…

nishanthmahasamudram:
…Not-eligible-and-Not-suitable-forms


AledWEvans (dismissed):
We should be carful not to put code into areas it dose not belong! 
The Question code should only be for data related to the question, how we display the results of the question in a document should be the concern of the template code in the PDF/Document service.

Use the question name to determin what text to put into the document in the PDF/Document service

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/PdfService.kt
```diff
@@ -74,6 +75,9 @@ class PdfService(
       DocumentSubjectType.OFFENDER_ADDRESS_CHECKS_FORM,
       DocumentSubjectType.OFFENDER_OPT_OUT_FORM,
       DocumentSubjectType.OFFENDER_NOT_ELIGIBLE_FORM,
+      -> {
+        data["documentFailureReason"] = caseView.documentFailureReason.firstOrNull()
```
andrewrlee, line 79: Should this be filtered to be an eligibility reason?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/PdfService.kt
```diff
@@ -89,7 +93,7 @@ class PdfService(
       DocumentSubjectType.OFFENDER_REFUSED_FORM,
       DocumentSubjectType.OFFENDER_NOT_SUITABLE_FORM,
       -> {
-        // nothing yet, add any form specific data here
+        data["documentFailureReason"] = caseView.documentFailureReason.firstOrNull()
```
andrewrlee, line 96: Should this be filtered to be an suitability reason?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/EligibilityAndSuitabilityCaseView.kt
```diff
@@ -117,6 +119,8 @@ data class SuitabilityCriterionProgress(
 data class Question(
   @Schema(description = "The question that is posed to the user", example = "a question...")
   val text: String,
+  @Schema(description = "The reason for failure", example = "a reason")
+  val documentFailureReason: String,
```
andrewrlee, line 123: does it need to be part of the question response?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/PdfService.kt
nishanthmahasamudram: We are doing it here `documentFailureReason = eligibility.getIneligibleReasons() + suitability.getUnsuitableReasons(),` So we will get eligibility failure reasons if any else suitability

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/PdfService.kt
nishanthmahasamudram: We are doing it here documentFailureReason = eligibility.getIneligibleReasons() + suitability.getUnsuitableReasons(), So we will get eligibility failure reasons if any else suitability

AledWEvans (approved):
LGTM

PR #177, opened 2025-04-23
CVSL-2779: com team caseload

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/mapper/AssessmentToAssessmentOverviewSummaryMapper.kt
```diff
@@ -39,7 +39,7 @@ class AssessmentToAssessmentOverviewSummaryMapper {
       location = prisonName,
       status = currentAssessment.status,
       responsibleCom = currentAssessment.responsibleCom?.toSummary(),
-      team = currentAssessment.team,
+      team = currentAssessment.teamCode,
```
andrewrlee, line 42: Guess it'd be good to renamed the representation version too.. looks like we only display it on the support view

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/mapper/AssessmentToAssessmentOverviewSummaryMapper.kt
justin-gardiner: Yes it would, I'll add a TODO locally so I remember to do it on next PR.

PR #178, opened 2025-04-23
CVSL-2743-AfER-PP-tasks-and-handover-rules

nishanthmahasamudram:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/ResidentialChecksServiceTest.kt
```diff
@@ -118,7 +124,7 @@ class ResidentialChecksServiceTest {
 
     whenever(assessmentService.getCurrentAssessment(PRISON_NUMBER)).thenReturn(assessmentEntity)
     whenever(addressService.getCurfewAddressCheckRequest(ADDRESS_REQUEST_ID, PRISON_NUMBER)).thenReturn(
-      aStandardAddressCheckRequest(),
+      aStandardAddressCheckRequest(true, true, true, true, true, true),
```
andrewrlee, line 127: loads of booleans like this are difficult to understand.. you could add an override in the test data for `aStandardAddressCheckRequestWithAllChecksComplete`? 

And then places where they aren't all true it'd be better to at least use named parameters

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/ResidentialChecksService.kt
```diff
@@ -148,20 +152,16 @@ class ResidentialChecksService(
...
 
-    return when {
-      suitable -> ResidentialChecksStatus.SUITABLE
-      unsuitable -> ResidentialChecksStatus.UNSUITABLE
-      inProgress -> ResidentialChecksStatus.IN_PROGRESS
-      else -> ResidentialChecksStatus.NOT_STARTED
-    }
+  fun getAddressesCheckStatus(addressCheckRequests: List<CurfewAddressCheckRequest>): ResidentialChecksStatus = when {
+    addressCheckRequests.any { isAddressSuitable(it.taskAnswers) } -> ResidentialChecksStatus.SUITABLE
+    addressCheckRequests.any { it.taskAnswers.any { answer -> answer.toTaskStatus() == UNSUITABLE } } -> ResidentialChecksStatus.UNSUITABLE
```
andrewrlee, line 162: think this should be "if All are unsuitable then the overall status is UNSUITABLE"

nishanthmahasamudram: No, if any of the status inside any addressCheckRequest is Unsuitable it should be UNSUITABLE.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/ResidentialChecksService.kt
```diff
@@ -148,20 +152,16 @@ class ResidentialChecksService(
...
-  private fun getAddressCheckStatus(addressCheckRequest: CurfewAddressCheckRequest): ResidentialChecksStatus {
-    val taskAnswers = addressCheckRequest.taskAnswers
-    val taskStatus = getPolicyTaskCodes().map { taskAnswers.find { answer -> answer.taskCode == it }.toTaskStatus() }
-
-    val unsuitable = taskStatus.any { it == UNSUITABLE }
-    val suitable = taskStatus.all { it == SUITABLE }
-    val inProgress = taskStatus.any { it == SUITABLE }
+  private fun isAddressSuitable(taskAnswers: MutableSet<ResidentialChecksTaskAnswer>): Boolean {
+    val taskStatus = getPolicyTaskCodes().map { taskAnswers.find { answer -> answer.taskCode == it }?.toTaskStatus() }
+    return taskStatus.all { it == SUITABLE }
```
andrewrlee, line 157: would change this to return a status rather than a boolean:

```
fun CurfewAddressCheckRequest.getStatus()
```
 then:
```kt
when {
    addressCheckRequests.any { it.status === SUITABLE } -> ResidentialChecksStatus.SUITABLE
    addressCheckRequests.all { it.status === UNSUITABLE } -> ResidentialChecksStatus.UNSUITABLE
    addressCheckRequests.any { it.status === IN_PROGRESS } -> ResidentialChecksStatus.IN_PROGRESS
   else -> ResidentialChecksStatus.NOT_STARTED


```

src/main/resources/migration/common/V56__add_address_check_status.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE assessment ADD COLUMN address_checks_status varchar(64);
+
+ALTER TABLE assessment
+    ADD CONSTRAINT assessment_address_checks_status_chk CHECK ( address_checks_status IN (
+                                                                              'NOT_STARTED',
+                                                                              'IN_PROGRESS',
+                                                                              'UNSUITABLE',
+                                                                              'SUITABLE'
+        ));
```
AledWEvans, line 9: I do wonder if we need to do the above DB enum as we have define it in the JPA
```
  @NotNull
  @Enumerated(EnumType.STRING)
  var addressChecksStatus: ResidentialChecksStatus? = null,
```
I suppose this still prevents it from being inserted via SQL, it's a pain though to update this when the enum is changed

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Assessment.kt
```diff
@@ -80,6 +81,10 @@ data class Assessment(
   @Column(name = "address_checks_complete")
   var addressChecksComplete: Boolean = false,
```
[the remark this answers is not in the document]
nishanthmahasamudram: Yup, it is been used in the frontend.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Assessment.kt
```diff
@@ -80,6 +81,10 @@ data class Assessment(
...
 
+  @NotNull
+  @Enumerated(EnumType.STRING)
+  var addressChecksStatus: ResidentialChecksStatus? = null,
```
AledWEvans, line 86: I don't think this needs to be null right!

`var addressChecksStatus: ResidentialChecksStatus = NOT_STARTED,`

AledWEvans (changes requested):
some comments

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Assessment.kt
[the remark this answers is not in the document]
nishanthmahasamudram: In the support view page to show whether Address Checks Complete.
        {
            key: { text: "Address Checks Complete" },
            value: { html: params.addressChecksComplete}
        },

src/main/resources/migration/common/V56__add_address_check_status.sql
nishanthmahasamudram: The @NotNull and @Enumerated(EnumType.STRING) annotations ensure that addressChecksStatus must have a non-null value and be stored as a string representation of the ResidentialChecksStatus enum. However, these annotations alone do not prevent users from inserting values other than those defined in the ResidentialChecksStatus enum.

To enforce that only specific enum values can be inserted, you would indeed need to add a constraint at the database level. The ALTER TABLE statement you provided is a good approach to achieve this.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/Assessment.kt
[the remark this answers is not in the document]
andrewrlee: I think we can remove this from the frontend and replace it with the status checks
Doesn't need to be done as this PR though

andrewrlee (approved):
LGTM

AledWEvans (approved):
LGTM

PR #180, opened 2025-04-24
CVSL-2779: Use a join on offender when fetching a potentially large n…

justin-gardiner:
…umber of assessments instead of selecting each offender for each assessment.


AledWEvans (approved):
LGTM

PR #187, opened 2025-04-30
CVSL-2819 Refactoring task config

andrewrlee:
All the logic for whether a given task is displayed or not and it's progress is now in one place rather than configured per status


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/state/Task.kt
```diff
@@ -0,0 +1,184 @@
...
+      .groupBy({ (role) -> role }, { (_, task) -> task })
+  }
+}
+
+enum class TaskStatus {
+  LOCKED,
+  READY_TO_START,
+  IN_PROGRESS,
+  COMPLETE,
+}
```
andrewrlee, line 184: This file was mostly generated from the code that sat in AssessmentStatus by this dodgy number:
```kt

fun main() {
    AssessmentStatus.values().flatMap { it.tasks().map { task -> it.name to task.key to task.value } }
        .flatMap { (code, tasks) ->
            tasks.map { task ->
                val (status, role) = code
                val taskStatus = if (task is Fixed) task.theStatus else "<dynamic>"
                listOf(task.task.name, role, status, taskStatus)
            }
        }.groupBy { it[0] }.forEach { (k, v) ->
            val roles = v.map { t -> t[1] }.distinct().joinToString(", ")
            val statuses = v.map { it.drop(2) }.groupBy({ it.last() }, { it.first() })
            println()

            println(
                "    ${k} {\n" +
                        "        override fun visibleForRoles() = setOf(${roles})\n" +
                        "        override fun taskStatus(assessment: Assessment): TaskStatus? {\n" +
                        "\t\t\twhen (assessment.status) {\n\t\t\t\t" +
                        statuses.map { (taskStatus, assessmentStatus) -> "${assessmentStatus.joinToString(", ")} -> return TaskStatus.${taskStatus}" }
                            .joinToString("\n\t\t\t\t") +
                        "\n\t\t\t\telse -> return null\n" +
                        "\t\t\t}\n" +
                        "        }\n" +
                        "    },"
            )

        }

```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/state/Task.kt
```diff
@@ -0,0 +1,184 @@
...
+
+enum class Task {
+
+  ASSESS_ELIGIBILITY {
+    override fun visibleForRoles() = setOf(PRISON_CA)
+    override fun taskStatus(assessment: Assessment): TaskStatus? = when (assessment.eligibilityChecksStatus) {
+      EligibilityStatus.NOT_STARTED -> READY_TO_START
+      EligibilityStatus.IN_PROGRESS -> IN_PROGRESS
+      else -> COMPLETE
+    }
```
andrewrlee, line 35: The logic for all tasks has been maintained apart from this guy.
I've moved this to read from the new persisted eligibility status instead of relying on overall assessment status

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/state/Task.kt
```diff
@@ -0,0 +1,184 @@
...
+
+  ENTER_CURFEW_ADDRESS {
+    override fun visibleForRoles() = setOf(PRISON_CA)
+    override fun taskStatus(assessment: Assessment): TaskStatus? = when (assessment.status) {
+      NOT_STARTED, ELIGIBILITY_AND_SUITABILITY_IN_PROGRESS, INELIGIBLE_OR_UNSUITABLE -> LOCKED
+      ELIGIBLE_AND_SUITABLE -> READY_TO_START
+      AWAITING_ADDRESS_AND_RISK_CHECKS, ADDRESS_AND_RISK_CHECKS_IN_PROGRESS -> COMPLETE
+      OPTED_OUT -> IN_PROGRESS
+      else -> null
+    }
```
andrewrlee, line 46: We'll likely gradually start to move these over as required to be more based on the state of an assessment rather than the AssessmentStatus.

This is to cope with things like if a case goes past this state, we still need to show progress rather than the case being locked

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/state/Task.kt
```diff
@@ -0,0 +1,184 @@
...
+
+  SEND_CHECKS_TO_PRISON {
+    override fun visibleForRoles() = setOf(PROBATION_COM)
+    override fun taskStatus(assessment: Assessment): TaskStatus? {
+      if (!assessment.status.isOneOf(AWAITING_ADDRESS_AND_RISK_CHECKS, ADDRESS_AND_RISK_CHECKS_IN_PROGRESS)) {
+        return null
+      }
+      return when {
+        assessment.addressChecksStatus == ResidentialChecksStatus.UNSUITABLE -> READY_TO_START
+        assessment.addressChecksStatus == ResidentialChecksStatus.SUITABLE && assessment.victimContactSchemeOptedIn != null && assessment.hasNonDisclosableInformation != null -> READY_TO_START
```
andrewrlee, line 128: previously this logic had to be shared across multiple states, now it only needs to be specified once

src/test/resources/test_data/responses/eligibility-and-suitability-view.json
```diff
@@ -16,7 +16,7 @@
       "PRISON_CA": [
         {
           "name": "ASSESS_ELIGIBILITY",
-          "progress": "READY_TO_START"
+          "progress": "IN_PROGRESS"
```
andrewrlee, line 19: Now showing correct state

src/test/resources/test_data/responses/eligibility-and-suitability-view.json
```diff
@@ -38,14 +38,16 @@
           "name": "PRINT_LICENCE",
           "progress": "LOCKED"
         }
-      ]
+      ],
+      "PROBATION_COM": [],
+      "PRISON_DM": []
```
andrewrlee, line 43: return empty rather than null for each role

src/test/resources/test_data/responses/eligibility-criterion-view.json
```diff
@@ -16,7 +16,7 @@
       "PRISON_CA": [
         {
           "name": "ASSESS_ELIGIBILITY",
-          "progress": "READY_TO_START"
+          "progress": "IN_PROGRESS"
```
andrewrlee, line 19: now showing correct status

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/state/Task.kt
```diff
@@ -0,0 +1,184 @@
...
+  },
+
+  CONSULT_THE_VLO_AND_POM {
+    override fun visibleForRoles() = setOf(PROBATION_COM)
+    override fun taskStatus(assessment: Assessment): TaskStatus? = when (assessment.status) {
+      AWAITING_ADDRESS_AND_RISK_CHECKS ->
+        if (assessment.victimContactSchemeOptedIn != null) COMPLETE else READY_TO_START
+
+      ADDRESS_AND_RISK_CHECKS_IN_PROGRESS -> COMPLETE
+      else -> null
```
andrewrlee, line 89: this could be simplified to: 
```kt
if (assessment.victimContactSchemeOptedIn != null) {
   return COMPLETE
} 
if (assessmentStatus == AWAITING_ADDRESS_AND_RISK_CHECKS){
   return NOT_STARTED
}
return null
```

AledWEvans: I like the when TBO

AledWEvans (approved):
LGTM
Looks like some good refactoring here.

PR #189, opened 2025-04-30
This test seems to be failing intermittently for me:

AledWEvans:
org.opentest4j.AssertionFailedError:
    expected: 2L
     but was: 1L
        at app//uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.integration.SupportResourceIntTest$GetAssessments.should get assessments when prison number given(SupportResourceIntTest.kt:382)

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/helpers/TestAssessmentRepository.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/helpers/TestAssessmentRepository.kt
@@ -8,5 +8,5 @@ import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.Offende
 @Repository
 interface TestAssessmentRepository : JpaRepository<Assessment, Long> {
   fun findByOffender(offender: Offender): List<Assessment>
-  fun findByOffenderPrisonNumber(prisonNumber: String): List<Assessment>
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/helpers/TestAssessmentRepository.kt
```diff
@@ -8,5 +8,5 @@ import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.entity.Offende
 @Repository
 interface TestAssessmentRepository : JpaRepository<Assessment, Long> {
   fun findByOffender(offender: Offender): List<Assessment>
-  fun findByOffenderPrisonNumber(prisonNumber: String): List<Assessment>
+  fun findByOffenderPrisonNumberOrderById(prisonNumber: String): List<Assessment>
```
andrewrlee, line 11: didn't know this existed 👍

andrewrlee (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/helpers/TestAssessmentRepository.kt
AledWEvans: Yes I added it, under the logic of if the repo code is for testing only should not be in production code. It's a nice way of separating test from prod code.. otherwise we end up with a lot of methods but only some used in prod code.

PR #191, opened 2025-05-01
CVSL-2772: Fix JPA annotation so assessment events are saved. Remove …

justin-gardiner:
…agent parameter for opt-in endpoint as now passed via headers.


andrewrlee (approved):
LGTM

PR #192, opened 2025-05-02
CVSL-2798 AfER | Road to production | Retrieve audit events for support

AledWEvans:
Server side code for support assessment events


gurps-singh (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/SupportResourceIntTest.kt
```diff
@@ -476,6 +490,135 @@ class SupportResourceIntTest : SqsIntegrationTestBase() {
...
+
+      val lastAssessmentEvent = assessmentEventList.last()
+      assertThat(lastAssessmentEvent.eventType).isEqualTo(AssessmentEventType.STATUS_CHANGE)
+      assertThat(lastAssessmentEvent.eventTime).isEqualTo(LocalDateTime.of(2024, 10, 11, 17, 19, 8))
+      assertThat(lastAssessmentEvent.username).isEqualTo("a-probation-user")
+      assertThat(lastAssessmentEvent.role).isEqualTo(UserRole.PROBATION_COM)
+      assertThat(lastAssessmentEvent.changes).contains("NOT_STARTED")
+      assertThat(lastAssessmentEvent.fullName).isEqualTo("Ceri Evans")
+      assertThat(lastAssessmentEvent.onBehalfOf).isEqualTo("BRI")
+      assertThat(lastAssessmentEvent.summary).isEqualTo("status changed from: 'NOT_STARTED', to: 'ELIGIBILITY_AND_SUITABILITY_IN_PROGRESS'")
```
andrewrlee, line 567: meh.. I like strictly matching against json response files to snapshot test these.. makes it much easier to see the response structure and detect backwards incompatible changes and keeps the tests nice and neat. 
It's also much easier to maintain these tests and just overwrite the old response with the new response, and eyeball the diff that an expected response file has changed in an expected way rather update code in potentially lots of places. 

A good example here is having to rely on a separate test representation - if this diverges from the actual json that's produced it'll be a pain to have to work out what the json response has changed to, then update the representation and then update the assertions.

PR #194, opened 2025-05-02
CVSL-2764-AFER-Soft-delete-address-check-requests

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AddressDeletionEvent.kt
```diff
@@ -0,0 +1,33 @@
...
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  @NotNull
+  val id: Long = -1,
+
+  @Enumerated(EnumType.STRING)
+  var addressDeleteReasonType: AddressDeleteReasonType? = null,
+
+  var addressDeleteOtherReason: String? = null,
+
```
nishanthmahasamudram, line 29: There is no need for bi-directional relation with CurfewAddressCheckRequest

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/service/AddressService.kt
```diff
@@ -235,6 +237,34 @@ class AddressService(
     curfewAddressCheckRequestRepository.save(curfewAddressCheckRequest)
   }
 
+  @Transactional
+  fun addressDeleteReason(
```
andrewrlee, line 241: `deleteAddress`?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/AddressResourceIntTest.kt
```diff
@@ -634,6 +637,79 @@ class AddressResourceIntTest : SqsIntegrationTestBase() {
...
+        .uri(ADDRESS_DELETE_REASON)
+        .headers(setAuthorisation(roles = listOf("ASSESS_FOR_EARLY_RELEASE_ADMIN"), agent = PRISON_CA_AGENT))
+        .bodyValue(anAddressDeleteReasonRequest())
+        .exchange()
+        .expectStatus()
+        .isCreated
+
+      val addressCheckRequest = curfewAddressCheckRequestRepository.findByIdOrNull(ADDRESS_REQUEST_ID)
+      val addressDeletionEvent = addressDeletionEventRepository.findByIdOrNull(addressCheckRequest?.addressDeletionEvent?.id)
+      assertThat(addressCheckRequest?.addressDeletionEvent).isEqualTo(addressDeletionEvent)
```
andrewrlee, line 690: think it'd be better to explicitly test the specific fields have been set on the version set on `addressCheckRequest` rather than make two repository calls.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AddressDeletionEvent.kt
```diff
@@ -0,0 +1,33 @@
...
+  @OneToOne(cascade = [CascadeType.ALL])
+  @JoinColumn(name = "assessment_event_id", referencedColumnName = "id")
+  val assessmentEvent: AssessmentEvent? = null,
```
andrewrlee, line 32: Should we override equals and hashcode?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AddressDeletionEvent.kt
AledWEvans, line 29: Just a suggestion, the normal convention would be just to call these fields :
AddressDeletionEvent :
          reasonType
          otherReason
Because the class already defines that it's about an address deletion event. you don't have to repeat it

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/AddressDeleteReason.kt
```diff
@@ -0,0 +1,34 @@
...
+  val addressDeleteReasonType: AddressDeleteReasonType,
+
+  @Schema(description = "Other reason to delete address", example = "Give details of information regards delete address")
+  val addressDeleteOtherReason: String?,
+)
+
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FILE)
+@Retention(AnnotationRetention.RUNTIME)
+@Constraint(validatedBy = [AddressDeleteReasonValidator::class])
+annotation class ValidAddressDeleteReason(
```
AledWEvans, line 26: In the past I would put annotations in there own package, something like model.valtiation.annotations,
Looks like we put it in with model kt file, which is fine until you use them in two dto's

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/AddressDeleteReason.kt
```diff
@@ -0,0 +1,34 @@
...
+import jakarta.validation.ConstraintValidator
+import jakarta.validation.ConstraintValidatorContext
+import jakarta.validation.Payload
+import jakarta.validation.constraints.NotNull
+import uk.gov.justice.digital.hmpps.hmppsassessforearlyreleaseapi.model.enum.AddressDeleteReasonType
+import kotlin.reflect.KClass
+
+@ValidAddressDeleteReason
+@Schema(description = "Records an offender's non disclosable information")
+data class AddressDeleteReason(
```
AledWEvans, line 14: are we post fixing these with request now? `AddressDeleteReasonRequest`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/resource/AddressResource.kt
```diff
@@ -500,4 +501,62 @@ class AddressResource(private val addressService: AddressService, private val ag
...
+        content = [
+          Content(
+            mediaType = "application/json",
+            schema = Schema(implementation = ErrorResponse::class),
+          ),
+        ],
+      ),
+    ],
+  )
+  fun addressDeleteReason(
```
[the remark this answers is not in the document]
AledWEvans: If we use this naming we should also change the naming of the other soft delete to match!?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/AddressResourceIntTest.kt
AledWEvans: +1

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/AddressResourceIntTest.kt
```diff
@@ -634,6 +637,79 @@ class AddressResourceIntTest : SqsIntegrationTestBase() {
...
+    )
+    @Test
+    fun `should update curfew address with address deletion reason`() {
+      webTestClient.post()
+        .uri(ADDRESS_DELETE_REASON)
+        .headers(setAuthorisation(roles = listOf("ASSESS_FOR_EARLY_RELEASE_ADMIN"), agent = PRISON_CA_AGENT))
+        .bodyValue(anAddressDeleteReasonRequest())
+        .exchange()
+        .expectStatus()
+        .isCreated
```
AledWEvans, line 686: is Created the correct status? 204 No Content may be?

AledWEvans (changes requested):
Some comments and changes

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/AddressResourceIntTest.kt
nishanthmahasamudram: Yes, we are just persisting the reason and navigating to other page.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AddressDeletionEvent.kt
nishanthmahasamudram: I wanted to maintain consistency across, so I am using addressDeleteReasonType for AddressDeletionEvent and few other places, I named it like this.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/integration/AddressResourceIntTest.kt
AledWEvans: Your not creating your updating and so it needs to be a 200 or 204,... 204 can also be for delete so I would go for 204!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/AddressDeleteReason.kt
AledWEvans: I thought this was the convention we are using from now on..

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AddressDeletionEvent.kt
AledWEvans: This is not how it should be done! this is a standard approach with OOP and DB design.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/model/AddressDeleteReason.kt
nishanthmahasamudram: No if it is used only for Request body then we can use but it is used in multiple places.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AddressDeletionEvent.kt
AledWEvans: As discussed changed this on the entity, leave as is on the DTO as dose not have context

AledWEvans (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AddressDeletionEvent.kt
```diff
@@ -0,0 +1,49 @@
...
+) {
+  override fun toString(): String = "AddressDeletionEvent(" +
+    "id=$id, " +
+    "addressDeleteReasonType=$reasonType, " +
+    "addressDeleteOtherReason=$otherReason, " +
+    "assessmentEvent=${assessmentEvent?.id}" +
```
andrewrlee, line 38: probably should update this too

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsassessforearlyreleaseapi/entity/AddressDeletionEvent.kt
```diff
@@ -0,0 +1,49 @@
...
+    ")"
+
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (other !is AddressDeletionEvent) return false
```
andrewrlee, line 43: think this should check for hibernate proxies: `if (other == null || Hibernate.getClass(this) != Hibernate.getClass(other)) return false`

andrewrlee (approved):
LGTM

